PR #1, opened 2022-09-07
updated readme

HamdanSCode:
updated readme with some new stuff

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,8 @@
-test
+# test this should be the largest heading
+## this should be the second largest
+###### this should be small af
+the general purpose o this readme right now is for testing purposes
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1 +1,8 @@
-test
\ No newline at end of file
+# test this should be the largest heading
```
ForzenAcademy: That's great that you're testing, but make this an official readme that describes that you will be making a drawing app. You can even give it a name!

HamdanSCode: Will do serji

HamdanSCode:
updated the files

ForzenAcademy (approved):

PR #4, opened 2022-09-24
FA 20 Updated Completion of the layer dialog

HamdanSCode:
I made a new thing! :)


app/src/main/java/com/example/drawingApp/DrawingFieldView.kt
```diff
@@ -20,6 +20,7 @@ class DrawingFieldView @JvmOverloads constructor(
     defStyle: Int = 0
 ) : View(context, attrs, defStyle) {
 
+    var onFinishLine: ((Bitmap) -> Unit)? = null
```
ForzenAcademy: Kdoc. What is this for

app/src/main/java/com/example/drawingApp/DrawingFieldView.kt
```diff
@@ -31,6 +32,7 @@ class DrawingFieldView @JvmOverloads constructor(
     private var path = Path()
     private var calcCanvas: Canvas? = null
```
ForzenAcademy: Better name. What is this for

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,23 +1,114 @@
...
         val colorCircleView = findViewById<SelectedColorView>(R.id.colorCircle)
         val drawingFieldView = findViewById<DrawingFieldView>(R.id.drawField)
 
+        drawingFieldView.onFinishLine = {
+            model.activeBitmap = it
+        }
+        findViewById<TextView>(R.id.newLayer).setOnClickListener {
+            showLayerAlertDialog()
+        }
+        if (model.activeBitmap != null) {
```
ForzenAcademy: Why do you need a null check AND a safe let?

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,23 +1,114 @@
...
+            model.activeBitmap?.let { it -> drawingFieldView.setBitmap(it) }
```
ForzenAcademy: You never need to specify `it` as a lambda name

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,23 +1,114 @@
...
+            drawingFieldView.invalidate()
+        }
+
         colorCircleView.onColorChange = { stateColor ->
             model.primaryColor = stateColor
             drawingFieldView.updateColor(stateColor)
         }
 
+        if (model.dialogOpen) showLayerAlertDialog()
+        //model.viewModelScope
```
ForzenAcademy: No stray comments

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,23 +1,114 @@
...
+    }
+
+    fun showLayerAlertDialog() {
+        val inflater = LayoutInflater.from(this@MainActivity)
+        val view = inflater.inflate(R.layout.layer_dialog, null)
+        val layerField = view.findViewById<EditText>(R.id.newLayerField)
+        val errorText = view.findViewById<TextView>(R.id.dialogCharCountError)
+
+        //if the layer number is not in the list already then set it to the field
+        if (model.layerNames.indexOf("Layer ${model.layerCounter}") == -1) {
```
ForzenAcademy: If you're not going to use the indexOf index, use `contains`

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,23 +1,114 @@
...
+            layerField.setText(getString(R.string.new_layer_hint, model.layerCounter))
+        } else {
+            //else if it is in the list, keep looping till we find that its not
+            while (model.layerNames.indexOf("Layer ${model.layerCounter}") != -1) {
```
ForzenAcademy: Again, you can use `contains`

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,23 +1,114 @@
...
+        }.create()
+
+        view.findViewById<TextView>(R.id.newLayerCancel).setOnClickListener {
+            dialog.dismiss()
+        }
+        view.findViewById<TextView>(R.id.newLayerOk).setOnClickListener {
+            errorText.visibility = View.INVISIBLE
+            if (layerField.text.toString() ==
+                getString(R.string.new_layer_hint, model.layerCounter)
+            ) model.layerCounter += 1
```
ForzenAcademy: brackets around this

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,23 +1,114 @@
...
+            model.layerNames.add(layerField.text.toString())
+            Toast.makeText(this@MainActivity, layerField.text.toString(), Toast.LENGTH_SHORT)
+                .show()
+            dialog.dismiss()
+        }
+        if (model.dialogOpen) {
```
ForzenAcademy: This feels very disjointed. What is this for?

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,23 +1,114 @@
...
+            layerField.setText(model.dialogText)
+        }
+        model.dialogOpen = true
```
ForzenAcademy: Booleans should be formed as questions, `isDialogOpen`

app/src/main/java/com/example/drawingApp/MainViewModel.kt
```diff
@@ -10,4 +11,13 @@ class MainViewModel : ViewModel() {
     @ColorInt
     var primaryColor = Color.BLACK
 
+    val layerNames: MutableList<String> = mutableListOf()
+    var layerCounter = 1
```
ForzenAcademy: I don't like the implementation of this saved in a counter. For one, you don't need to store this to be able to calculate it. You can always just count up until you find one you don't have. And what if they delete one later? What if I save this project for reopening later? I have to now store a layerCounter? It's making a dependency on this field that it probably doesn't need

app/src/main/java/com/example/drawingApp/MainViewModel.kt
```diff
@@ -10,4 +11,13 @@ class MainViewModel : ViewModel() {
...
+
+    var activeBitmap: Bitmap? = null
+
+    var dialogOpen = false
+    var dialogText = ""
```
ForzenAcademy: More descriptive. `usersCurrentDialogInput`?

app/src/main/java/com/example/drawingApp/DrawingFieldView.kt
HamdanSCode: added Kdoc

app/src/main/java/com/example/drawingApp/DrawingFieldView.kt
HamdanSCode: renamed to drawingCanvas

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: made it just the let

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: removed the given name

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: removed

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: swapped to contains, I really was just doing a longer version of contains

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: removed

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: I'm not sure what you mean, do you mean around the model.layerCounter?

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: This is what allows it to know to set the text / to maintain the dialog being open, the first time in the concrete is to call the func to open the dialog if it was open previously, in the function it checks if the dialog was previously open and if it was it knows to grab the dialogue text in the view model

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: renamed

app/src/main/java/com/example/drawingApp/MainViewModel.kt
HamdanSCode: your right about it creating the dependency and the future issue with delete, instead it will now search for the first available layer name starting from 1. Removed some redundant code with the removal of model.layerCounter

app/src/main/java/com/example/drawingApp/MainViewModel.kt
HamdanSCode: renamed to currentDialogText

app/src/main/java/com/example/drawingApp/DrawingFieldView.kt
```diff
@@ -20,6 +20,11 @@ class DrawingFieldView @JvmOverloads constructor(
     defStyle: Int = 0
 ) : View(context, attrs, defStyle) {
 
+    /**
+     * Whenever a line is finished being drawn call this
+     * specifically triggers on motion event up / cancel
+     */
+    var onFinishLine: ((Bitmap) -> Unit)? = null
```
ForzenAcademy: Ok but why

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,23 +1,111 @@
...
+import android.widget.EditText
+import android.widget.TextView
+import android.widget.Toast
 import androidx.activity.viewModels
+import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatActivity
+import androidx.core.widget.addTextChangedListener
 
 class MainActivity : AppCompatActivity() {
+    private val MAX_CHAR_LEN = 16
```
ForzenAcademy: Put constants like this in a companion object at the bottom

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,23 +1,111 @@
...
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_main)
-
+        enableStrictMode()
         val colorCircleView = findViewById<SelectedColorView>(R.id.colorCircle)
         val drawingFieldView = findViewById<DrawingFieldView>(R.id.drawField)
 
+        drawingFieldView.onFinishLine = {
+            model.activeBitmap = it
```
ForzenAcademy, line 27: What if they rotate or get a call halfway through a bigass line

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,23 +1,111 @@
...
+            showLayerAlertDialog()
+        }
+
+        model.activeBitmap?.let {
+            drawingFieldView.setBitmap(it)
+            drawingFieldView.invalidate()
+        }
+
+
+
```
ForzenAcademy: Extra space

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,23 +1,111 @@
...
+    fun showLayerAlertDialog() {
+        val inflater = LayoutInflater.from(this@MainActivity)
+        val view = inflater.inflate(R.layout.layer_dialog, null)
+        val layerField = view.findViewById<EditText>(R.id.newLayerField)
+        val errorText = view.findViewById<TextView>(R.id.dialogCharCountError)
+        var layerCounter = 1
+
+        //find the first available layer name not taken
+        while (model.layerNames.contains("Layer ${layerCounter}")) {
```
ForzenAcademy: String resource

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,23 +1,111 @@
...
+            layerCounter += 1
+        }
+        layerField.setText(getString(R.string.new_layer_hint, layerCounter))
+
+
+        //needs to be after the set text or else the textchanged listener will overwrite the previous text needed for rotations
+        layerField.addTextChangedListener {
+            model.currentDialogText = layerField.text.toString()
+            if (it.toString().trim().length > MAX_CHAR_LEN) {
+                errorText.visibility = View.VISIBLE
```
ForzenAcademy: `errorText.isVisible = it.toString().trim().length > MAX_CHAR_LEN`?

app/src/main/java/com/example/drawingApp/MainViewModel.kt
```diff
@@ -1,5 +1,6 @@
 package com.example.drawingApp
 
+import android.graphics.Bitmap
```
ForzenAcademy: `MainViewModel`? Why not something more related to Color or drawing or?

app/src/main/java/com/example/drawingApp/MainViewModel.kt
```diff
@@ -10,4 +11,12 @@ class MainViewModel : ViewModel() {
     @ColorInt
     var primaryColor = Color.BLACK
 
+    val layerNames: MutableList<String> = mutableListOf()
+
+    var activeBitmap: Bitmap? = null
+
+    var isDialogOpen = false
+    var currentDialogText = ""
```
ForzenAcademy: Which dialog for these? Name more clear

app/src/main/res/drawable/text_btn_background_.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle">
+    <corners
+        android:radius="14dp"/>
```
ForzenAcademy: dimen

ForzenAcademy (changes requested):

app/src/main/java/com/example/drawingApp/DrawingFieldView.kt
HamdanSCode: changed and explained

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: left a comment in the program as well

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: created and moved

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: changed name and changed location to the on move, avoiding the issue.

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: removed, formatting doesn't get rid of extra space I suppose.

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: updated

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: implemented but it feels far more awkward than the previous.

app/src/main/java/com/example/drawingApp/MainViewModel.kt
HamdanSCode: I had no context if the view model was going to be used for anything else so it had a temporary class name, updated to DrawingViewModel.

app/src/main/java/com/example/drawingApp/MainViewModel.kt
HamdanSCode: updated to isLayerCreationDialogOpen, and currentLayerCreationDialogText

app/src/main/res/drawable/text_btn_background_.xml
HamdanSCode: added corner_rounding1-6 to the dimens

app/src/main/java/com/example/drawingApp/MainActivity.kt
ForzenAcademy: Yes. You have an if statement here, let's put brackets around its contents

app/src/main/java/com/example/drawingApp/MainViewModel.kt
ForzenAcademy: *you're

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: implemented

app/src/main/java/com/example/drawingApp/MainActivity.kt
ForzenAcademy: No reason to use an if statement when you are already processing a boolean result

app/src/main/java/com/example/drawingApp/MainViewModel.kt
HamdanSCode: er'uoy

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,23 +1,118 @@
...
+        layerField.setText(getString(R.string.new_layer_hint, layerCounter))
+
+        //needs to be after the set text or else the textchanged listener will overwrite the previous text needed for rotations
+        layerField.addTextChangedListener {
+            model.currentLayerCreationDialogText = layerField.text.toString()
+            errorText.isVisible = it.toString().trim().length > MAX_CHAR_LEN
+            if (errorText.isVisible) {
+                layerField.setText(layerField.text.toString().dropLast(1))
+                //visibility needs to be after or else the dropLast will make it never visible
+                errorText.visibility = View.VISIBLE
```
ForzenAcademy: Why are you changing its visibility when you already set its visibility on line 64?

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,23 +1,118 @@
...
+            layerCounter += 1
+        }
+        layerField.setText(getString(R.string.new_layer_hint, layerCounter))
+
+        //needs to be after the set text or else the textchanged listener will overwrite the previous text needed for rotations
+        layerField.addTextChangedListener {
+            model.currentLayerCreationDialogText = layerField.text.toString()
+            errorText.isVisible = it.toString().trim().length > MAX_CHAR_LEN
+            if (errorText.isVisible) {
+                layerField.setText(layerField.text.toString().dropLast(1))
```
ForzenAcademy: I don't see why you need to do this at all.... what is this solving? If you really wanted to not let them type any more you could take the first MAX_CHAR_LEN substring. If I paste 30 characters your logic fails. Or just set max length on the edittext

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,23 +1,118 @@
...
+
+        val dialog = AlertDialog.Builder(this).apply {
+            setView(view)
+        }.create()
+
+        view.findViewById<TextView>(R.id.newLayerCancel).setOnClickListener {
+            dialog.dismiss()
+        }
+        view.findViewById<TextView>(R.id.newLayerOk).setOnClickListener {
+            errorText.visibility = View.INVISIBLE
```
ForzenAcademy: Shouldn't this check for duplicates? And in that case, why hide the errorText?

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: because of the droplast it will never show the error text without the visibility being set there

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: if you paste 30 characters it does still work correctly, however you are correct substring would be much cleaner, if they end up trying to paste an extremely long string it would create a massive recursive call.

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: it can do that, it wasn't specified so I hadn't done that, but I will implement a check.
The reason the error text was being hidden was because the text for the layer could only conform to 16 characters which means that it would've been valid in all cases without checking for a duplicate layer with the same name.

Will update with a check and slightly different logic.

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: now instead of a droplast it is a substing but same thought applies.

ForzenAcademy: This doesn't make sense. You are setting its visibility, using it in an if statement, and then setting it again using a different method. For one, errorText will ALWAYS be visible here. You are confirming so with your if statement. What part am I not seeing?

app/src/main/java/com/example/drawingApp/DrawingFieldView.kt
```diff
@@ -20,6 +20,14 @@ class DrawingFieldView @JvmOverloads constructor(
...
 ) : View(context, attrs, defStyle) {
 
+    /**
+     * Whenever a line is being drawn call this
+     * specifically triggers on motion event move
+     * Purpose is to update the model's bitmap whenever we are drawing on the bitmap
+     * to keep the state of the bitmap consistent in the case of a outside interference
+     * such as a phone call or a sudden orientation change
+     */
+    var onLineDrawing: ((Bitmap) -> Unit)? = null
```
ForzenAcademy: Let's make this `onBitmapUpdated` because it won't just be lines in the future

HamdanSCode: changed

ForzenAcademy (approved):

PR #5, opened 2022-09-27
FA-24, added reusable dialog via Object

HamdanSCode:


app/src/main/java/com/example/drawingApp/DialogObject.kt
```diff
@@ -0,0 +1,119 @@
...
+        model: DrawingViewModel,
+        onDialogSubmission: (String) -> Unit
+    ) {
+        val inflater = LayoutInflater.from(c)
+        val view = inflater.inflate(R.layout.layer_dialog, null)
+        val layerField = view.findViewById<EditText>(R.id.newLayerField)
+        val errorText = view.findViewById<TextView>(R.id.dialogError)
+        val layerNames = mutableListOf<String>()
+        for (x in model.layerNames) layerNames.add(x.text)
+        var layerCounter = 1
```
brosscode: should this should be a constant in a companion object?

app/src/main/java/com/example/drawingApp/DialogObject.kt
```diff
@@ -0,0 +1,119 @@
...
+                    R.string.new_layer_hint,
+                    layerCounter
+                )
+            )
+        ) {
+            layerCounter += 1
+        }
+        layerField.setText(c.resources.getString(R.string.new_layer_hint, layerCounter))
+
+        //needs to be after the set text or else the textchanged listener will overwrite the previous text needed for rotations
```
brosscode: comments left in code following push should describe what it does for someone who is glancing over the function, you did this above for the while loop but not here. 
Nic had me remove comments from all prod code, so if you feel something is hard to read afterwards, there might be spaces to reformat

app/src/main/java/com/example/drawingApp/DialogObject.kt
```diff
@@ -0,0 +1,119 @@
...
+import androidx.core.widget.addTextChangedListener
+import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
+import com.google.android.material.bottomsheet.BottomSheetDialog
+
+object DialogObject {
+
+    fun showLayerAlertDialog(
+        c: Context,
+        model: DrawingViewModel,
```
brosscode: I think you should consider renaming this to viewModel, having a model in here was a little confusing compared to calling it a viewModel

app/src/main/java/com/example/drawingApp/DialogObject.kt
```diff
@@ -0,0 +1,119 @@
...
+                )
+            )
+        ) {
+            layerCounter += 1
+        }
+        layerField.setText(c.resources.getString(R.string.new_layer_hint, layerCounter))
+
+        //needs to be after the set text or else the textchanged listener will overwrite the previous text needed for rotations
+        layerField.addTextChangedListener {
+            val fieldString = it.toString().trim()
```
brosscode: unsure why we used .trim() instead of just working with the string, Logic should flexible to handle a string rather than needing it trimmed

app/src/main/java/com/example/drawingApp/DialogObject.kt
```diff
@@ -0,0 +1,119 @@
...
+        }.create()
+
+        view.findViewById<TextView>(R.id.newLayerCancel).setOnClickListener {
+            dialog.dismiss()
+        }
+
+        view.findViewById<TextView>(R.id.newLayerOk).setOnClickListener {
+            val fieldString = layerField.text.toString().trim()
+            if (fieldString.length > MainActivity.MAX_CHAR_LEN) {
+                errorText.visibility = View.VISIBLE
```
brosscode: I would use isVisible instead of .visibility here

app/src/main/java/com/example/drawingApp/DialogObject.kt
```diff
@@ -0,0 +1,119 @@
...
+            } else if (layerNames.contains(fieldString)) {
+                errorText.setText(R.string.layerExistsError)
+                errorText.visibility = View.VISIBLE
```
brosscode: see above

app/src/main/java/com/example/drawingApp/DialogObject.kt
```diff
@@ -0,0 +1,119 @@
...
+            layerField.setText(model.currentLayerCreationDialogText)
+        }
+        model.isLayerCreationDialogOpen = true
+        dialog.setOnDismissListener { model.isLayerCreationDialogOpen = false }
+        dialog.show()
+    }
+
+    fun showEditLayerDialog(c: Context, model: DrawingViewModel) {
+        val data = model.layerNames
+        lateinit var myAdapter: EditLayerAdapter
```
brosscode: is this a lateinit because it needs to be? or was it possible to do this after you set the view variable? I need to do this still so I will come back to this

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -118,6 +66,6 @@ class MainActivity : AppCompatActivity() {
     }
 
     companion object {
-        private val MAX_CHAR_LEN = 16
+        val MAX_CHAR_LEN = 16
```
brosscode: I declared these as const vals

app/src/main/res/drawable/ic_baseline_delete_24.xml
```diff
@@ -0,0 +1,5 @@
+<vector android:height="24dp" android:tint="#000000"
+    android:viewportHeight="24" android:viewportWidth="24"
+    android:width="24dp" xmlns:android="http://schemas.android.com/apk/res/android">
+    <path android:fillColor="@android:color/white" android:pathData="M6,19c0,1.1 0.9,2 2,2h8c1.1,0 2,-0.9 2,-2V7H6v12zM19,4h-3.5l-1,-1h-5l-1,1H5v2h14V4z"/>
+</vector>
```
brosscode: Provide image of what this looks like

app/src/main/res/drawable/ic_baseline_edit_24.xml
```diff
@@ -0,0 +1,5 @@
+<vector android:height="24dp" android:tint="#000000"
+    android:viewportHeight="24" android:viewportWidth="24"
+    android:width="24dp" xmlns:android="http://schemas.android.com/apk/res/android">
+    <path android:fillColor="@android:color/white" android:pathData="M3,17.25V21h3.75L17.81,9.94l-3.75,-3.75L3,17.25zM20.71,7.04c0.39,-0.39 0.39,-1.02 0,-1.41l-2.34,-2.34c-0.39,-0.39 -1.02,-0.39 -1.41,0l-1.83,1.83 3.75,3.75 1.83,-1.83z"/>
+</vector>
```
brosscode: same as above. additionally, consider renaming these

app/src/main/java/com/example/drawingApp/DialogObject.kt
HamdanSCode: no, if it was a companion object then on deletions I would have to also count down.

HamdanSCode: and if a rename happens to a different name than layer x it interrupts the flow potentially and gives an incorrect output value for the hint

app/src/main/java/com/example/drawingApp/DialogObject.kt
HamdanSCode: because the trim is what we are actually going to be taking for the value, if its not valid in that form the user needs to know, no reason to do the hard work of rewriting a trim for it. The trim function really just sends back the string with no whitespace in front or behind it.

app/src/main/java/com/example/drawingApp/DialogObject.kt
HamdanSCode: yes, it needs to be a lateinit to reference the adapter in the lambdas.

app/src/main/java/com/example/drawingApp/DialogObject.kt
HamdanSCode: its kind of more of a warning, because it looks like it could be moved earlier.

app/src/main/java/com/example/drawingApp/DialogObject.kt
HamdanSCode: I think thats a fair point, incorporated.

app/src/main/java/com/example/drawingApp/DialogObject.kt
HamdanSCode: if the error was already visible from the text string input earlier it would then not show the error to user the isVisible

app/src/main/java/com/example/drawingApp/DialogObject.kt
HamdanSCode: because of the requirement of the below, it would be weird to use isVisible here and setting visibility there.

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: fair

app/src/main/res/drawable/ic_baseline_delete_24.xml
HamdanSCode: will do

app/src/main/res/drawable/ic_baseline_edit_24.xml
HamdanSCode: renamed

HamdanSCode:
edit
delete

The edit image and delete image used for buttons

app/src/main/java/com/example/drawingApp/DialogObject.kt
```diff
@@ -0,0 +1,127 @@
...
+import android.widget.TextView
+import android.widget.Toast
+import androidx.appcompat.app.AlertDialog
+import androidx.core.view.isVisible
+import androidx.core.widget.addTextChangedListener
+import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
+import com.google.android.material.bottomsheet.BottomSheetDialog
+
+object DialogObject {
```
ForzenAcademy: Is this a utility class? I think so. `DialogUtility`?

app/src/main/java/com/example/drawingApp/DialogObject.kt
```diff
@@ -0,0 +1,127 @@
...
+
+
+    fun showLayerAlertDialog(
```
ForzenAcademy: Kdoc this function

app/src/main/java/com/example/drawingApp/DialogObject.kt
```diff
@@ -0,0 +1,127 @@
...
+        c: Context,
+        viewModel: DrawingViewModel,
```
ForzenAcademy: I don't think you need this nor should it be tied to this. You could bring in a list of existing layer names, or even a lambda of `isValidName: (String) -> Boolean` and `onTextChanged` so you can save it. Let's make it more flexible

app/src/main/java/com/example/drawingApp/DialogObject.kt
```diff
@@ -0,0 +1,127 @@
...
+import androidx.core.widget.addTextChangedListener
+import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
+import com.google.android.material.bottomsheet.BottomSheetDialog
+
+object DialogObject {
+
+
+    fun showLayerAlertDialog(
+        c: Context,
```
ForzenAcademy: Let's name this `context` so it's not ambiguous later on

app/src/main/java/com/example/drawingApp/DialogObject.kt
```diff
@@ -0,0 +1,127 @@
...
+        }
+        layerField.setText(c.resources.getString(R.string.new_layer_hint, layerCounter))
+
+        //this text changed listener needs to be after the set text
+        //checks validity of the string length in the layer field, if not valid shows an error
+        layerField.addTextChangedListener {
+            val fieldString = it.toString().trim()
+            viewModel.currentLayerCreationDialogText = fieldString
+            errorText.setText(R.string.charCountError)
+            errorText.isVisible = fieldString.length > MainActivity.MAX_CHAR_LEN
```
ForzenAcademy: Does this constant still make sense to belong to Main Activity? The activity doesn't care how long the name is. Either the ViewModel cares or this dialog cares

app/src/main/java/com/example/drawingApp/DialogObject.kt
```diff
@@ -0,0 +1,127 @@
...
+        }.create()
+
+        view.findViewById<TextView>(R.id.newLayerCancel).setOnClickListener {
+            dialog.dismiss()
+        }
+
+        view.findViewById<TextView>(R.id.newLayerOk).setOnClickListener {
+            val fieldString = layerField.text.toString().trim()
+            if (fieldString.length > MainActivity.MAX_CHAR_LEN) {
+                errorText.visibility = View.VISIBLE
```
ForzenAcademy: No more using `.visibility`, it's obsolete with `.isVisible`

app/src/main/java/com/example/drawingApp/DialogObject.kt
```diff
@@ -0,0 +1,127 @@
...
+            } else if (layerNames.contains(fieldString)) {
+                errorText.setText(R.string.layerExistsError)
+                errorText.visibility = View.VISIBLE
+            } else {
+                layerNames.add(fieldString)
+                onDialogSubmission(fieldString)
+                //model.layerNames.add(LayerViewModel(fieldString))
```
ForzenAcademy: Clean up comments

app/src/main/java/com/example/drawingApp/DialogObject.kt
```diff
@@ -0,0 +1,127 @@
...
+                Toast.makeText(c, fieldString, Toast.LENGTH_SHORT)
+                    .show()
+                dialog.dismiss()
+            }
+        }
+        //this is here so that in the case that the dialogue was previously open
```
ForzenAcademy: `Dialog` is for software components. `Dialogue` is for conversational words in a story

app/src/main/java/com/example/drawingApp/DialogObject.kt
```diff
@@ -0,0 +1,127 @@
...
+        //we will immediately know to set the text to what it was previously
+        //the reason it is here, is so that we avoid overriding the correct text
+        if (viewModel.isLayerCreationDialogOpen) {
+            layerField.setText(viewModel.currentLayerCreationDialogText)
```
ForzenAcademy: You don't need this whole flow, just bring in a default text field with "" as its default

app/src/main/java/com/example/drawingApp/DialogObject.kt
```diff
@@ -0,0 +1,127 @@
...
+        }
+        viewModel.isLayerCreationDialogOpen = true
+        dialog.setOnDismissListener { viewModel.isLayerCreationDialogOpen = false }
+        dialog.show()
+    }
+
+    fun showEditLayerDialog(c: Context, viewModel: DrawingViewModel) {
```
ForzenAcademy: `context` and same not needed `viewModel` comments

ForzenAcademy: This is more of a dialog for _view_ing layers

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
```diff
@@ -11,7 +11,7 @@ class DrawingViewModel : ViewModel() {
     @ColorInt
     var primaryColor = Color.BLACK
 
-    val layerNames: MutableList<String> = mutableListOf()
+    val layerNames: MutableList<LayerViewModel> = mutableListOf()
```
ForzenAcademy: You're bleeding `ViewModel` outside the recyclerview use case. Just name this `Layer` if that's the case, and update the name to be `layers`. And letting this be public lets anyone you are passing it to have edit access to it, which seems dangerous

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -28,7 +22,12 @@ class MainActivity : AppCompatActivity() {
         }
 
         findViewById<TextView>(R.id.newLayer).setOnClickListener {
-            showLayerAlertDialog()
+            DialogObject.showLayerAlertDialog(this, model,
+                onDialogSubmission = { model.layerNames.add(LayerViewModel(it)) }
```
ForzenAcademy: This here shows a good pattern for this. Eliminate the need to the model as a dependency

app/src/main/java/com/example/drawingApp/editLayerAdapter.kt
```diff
@@ -0,0 +1,77 @@
+package com.example.drawingApp
```
ForzenAcademy: This file has a lower case letter at the start,  please fix

app/src/main/java/com/example/drawingApp/editLayerAdapter.kt
```diff
@@ -0,0 +1,77 @@
...
+     * determines what will be performed when clicked
+     * currently allows the editing of the layer at that position of data
+     */
+    private val onEditClick: (LayerViewModel) -> Unit,
+    /**
+     * called when the delete image is clicked in each view of the recycler view
+     * determines what will be performed when clicked
+     * currently deletes the layer at that position of data
+     */
+    private val onDeleteClick: (LayerViewModel) -> Unit
```
ForzenAcademy: Good! 🎉

app/src/main/java/com/example/drawingApp/editLayerAdapter.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    override fun onBindViewHolder(holder: ViewHolder, position: Int) {
+        holder.bind(data[position])
+    }
+
+    override fun getItemCount(): Int = data.size
+
+}
+
+class LayerViewModel(val text: String)
```
ForzenAcademy: Along with my previous comment, I think you need a `Layer` object for the concept of a layer, but here this makes sense to leave as `LayerViewModel` because this is merely _presenting_ rather than doing Layer logic, so this is fine to keep, but make a `Layer` object for uses outside the RecyclerView

app/src/main/java/com/example/drawingApp/editLayerAdapter.kt
```diff
@@ -0,0 +1,77 @@
...
+
+abstract class ViewHolder(view: View) : RecyclerView.ViewHolder(view) {
+    abstract fun bind(model: LayerViewModel)
+}
+
+class LayerViewHolder(
+    view: View,
+    edit: (LayerViewModel) -> Unit,
```
ForzenAcademy: Lambda names should start with `on`

app/src/main/java/com/example/drawingApp/editLayerAdapter.kt
```diff
@@ -0,0 +1,77 @@
...
+    delete: (LayerViewModel) -> Unit
+) :
+    ViewHolder(view) {
```
ForzenAcademy: This can be on previous line

app/src/main/java/com/example/drawingApp/editLayerAdapter.kt
```diff
@@ -0,0 +1,77 @@
...
+    private val textView: TextView
+    private val editView: ImageView
+    private val deleteView: ImageView
+    val editLayer = edit
```
ForzenAcademy: Not private?

app/src/main/res/values/strings.xml
```diff
@@ -1,6 +1,6 @@
 <resources>
     <string name="app_name">DrawingApp</string>
-    <string name="new_layer_dialog_title">Create New Layer</string>
+    <string name="new_layer_dialog_title">Create / Edit Layer</string>
```
ForzenAcademy: Your title for your dialog should be smart enough to know which function it is doing, pass it in to your function based on context

ForzenAcademy (dismissed):

app/src/main/java/com/example/drawingApp/DialogObject.kt
HamdanSCode: renamed

app/src/main/java/com/example/drawingApp/DialogObject.kt
HamdanSCode: renamed

app/src/main/java/com/example/drawingApp/DialogObject.kt
HamdanSCode: no it doesnt, moved to be a part of the utility object.

app/src/main/java/com/example/drawingApp/DialogObject.kt
HamdanSCode: adjusted, feels a bit awkward but its working.

app/src/main/java/com/example/drawingApp/DialogObject.kt
HamdanSCode: removed

app/src/main/java/com/example/drawingApp/DialogObject.kt
HamdanSCode: To clarify this, you mean that if they were writing a dialogue name, were interrupted or had a rotation occur, you want it to now have a empty string?

app/src/main/java/com/example/drawingApp/DialogObject.kt
HamdanSCode: renamed to showViewingLayersDialog

app/src/main/java/com/example/drawingApp/DialogObject.kt
HamdanSCode: IMPLEMENTED

app/src/main/java/com/example/drawingApp/editLayerAdapter.kt
HamdanSCode: Please revisit, on the next commit, I think a Layer object will definitely be needed down the line, but I changed back the layerNames from layerViewModels back to a list of strings.

app/src/main/java/com/example/drawingApp/editLayerAdapter.kt
HamdanSCode: renamed

app/src/main/java/com/example/drawingApp/editLayerAdapter.kt
HamdanSCode: moved

app/src/main/java/com/example/drawingApp/editLayerAdapter.kt
HamdanSCode: updated both to be private, no reason for them not to be.

app/src/main/res/values/strings.xml
HamdanSCode: updated to handle

app/src/main/java/com/example/drawingApp/DialogObject.kt
HamdanSCode: implemented

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
HamdanSCode: changed back to string, your right it is out of scope of ticket. handled everything with this as an string list. In the future it would probably need to be a layer list of Layer objects.

app/src/main/java/com/example/drawingApp/DialogUtility.kt
```diff
@@ -0,0 +1,241 @@
...
+    /**
+     * Shows a Dialog in the context it is given.
+     * The dialogs purpose is to allow the user to name a layer.
+     * the dialog should then work with that layer in the onDialogSubmission lambda
+     * ideally returning it to some location.
+     * context is the context to which you want to open the alert in.
+     * layerNames is the list of strings in which you want to check against for validity of the string you want to submit.
+     * refText is the text the dialog will use as a default if there was a screen interruption.
+     * wasOpen is what it uses to determine if there was a screen interruption while it was open.
+     * isCreating is for assigning title correctly, true is for creating a dialog, false is for editing.
```
brosscode: A lot of these names are difficult to follow, while explaining their purpose is nice, I think being a little more verbose in naming them is a tradeoff you should consider when creating them.

app/src/main/java/com/example/drawingApp/DialogUtility.kt
```diff
@@ -0,0 +1,241 @@
...
+        val layerField = view.findViewById<EditText>(R.id.newLayerField)
+        val errorText = view.findViewById<TextView>(R.id.dialogError)
+        val titleText = view.findViewById<TextView>(R.id.layerDialogTitle)
+        if (isCreating) titleText.setText(R.string.new_layer_dialog_title)
+        else titleText.setText(R.string.edit_layer_dialog_title)
+        var layerCounter = 1
+
+
+        //find the first available layer name not taken
+        while (layerNames.contains(
```
brosscode: I like this, very condense loop. Formatting makes it longer but not sure how much you can really do about that honestly

app/src/main/java/com/example/drawingApp/DialogUtility.kt
```diff
@@ -0,0 +1,241 @@
...
+        }.create()
+
+        view.findViewById<TextView>(R.id.newLayerCancel).setOnClickListener {
+            dialog.dismiss()
+        }
+
+        view.findViewById<TextView>(R.id.newLayerOk).setOnClickListener {
+            val fieldString = layerField.text.toString().trim()
+            errorText.isVisible = fieldString.length > MAX_CHAR_LEN
+            if (errorText.isVisible) {
```
brosscode: Could use a when block here, might be cleaner

app/src/main/java/com/example/drawingApp/DialogUtility.kt
```diff
@@ -0,0 +1,241 @@
...
+
+        lateinit var myAdapter: EditLayerAdapter
+        myAdapter = EditLayerAdapter(
+            data = data,
+            onDeleteClick = { layerModel ->
+                val index = data.indexOf(layerModel)
+                data.removeAt(index)
+                myAdapter.notifyItemRemoved(index)
+            },
+            onEditClick = { layerModel ->
```
brosscode, line 135: I did a lot of this logic managing the layerList in my viewModel, but I guess I didn't consider doing it in the custom class for this.

I don't know if this is "wrong", it just strikes me as a distinctly different choice. All of this is related to view level logic, and I see context being passed around a lot, so I don't know if reworking this into a viewModel would even be worthwhile at this point

Curious what Nic will have on this one

app/src/main/java/com/example/drawingApp/DialogUtility.kt
```diff
@@ -0,0 +1,241 @@
...
+        view.findViewById<RecyclerView>(R.id.editLayerRecycler).apply {
+            layoutManager = LinearLayoutManager(context)
+            adapter = myAdapter
+        }
+
+        sheet.setContentView(view)
+        sheet.setOnDismissListener { onDismiss() }
+        sheet.show()
+
+        if (wasAlertOpen && activeEditIndex != -1) {
```
brosscode: the activeEditIndex at -1 here strikes me as odd. I feel like you basically want a nullable here but are doing it in a different way?

yeah after looking at all the points around the code referencing this variable, I think it should be a nullable, this one does feel wrong

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
```diff
@@ -15,8 +15,16 @@ class DrawingViewModel : ViewModel() {
 
     var activeBitmap: Bitmap? = null
 
-    var isLayerCreationDialogOpen = false
-    var currentLayerCreationDialogText = ""
+    var isLayerCreationDialogOpen = false   //new Layer
+    var currentLayerDialogText = ""
```
brosscode: I like nullable declarations, I think working around the null makes your code safer in case anything in future were to change something and it went wrong, you would have robust code to handle the null in a way that doesn't crash the app.

But if we're honest here, the logic in this app won't be touched by others and its likely safe to include variables like this.

I hope.

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
```diff
@@ -15,8 +15,16 @@ class DrawingViewModel : ViewModel() {
...
 
+    var isLayerViewDialogOpen = false       //edit layer
+    var isLayerViewEditDialogOpen = false   //view inside edit layer
+    var layerViewEditDialogIndex:Int = LAYER_VIEW_EDIT_INDEX
+    var currentViewLayerDialogText = ""
+
+    companion object{
+        const val LAYER_VIEW_EDIT_INDEX = -1
```
brosscode: I like using the companion object, but I think it is interesting you default it to -1 here, but don't have similar constants for layerCounter from the dialog util

I was able to import my companion object in my viewModel into anywhere else I needed the constants, not sure if that would work for you.

More of an idea than a request for a change

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
```diff
@@ -15,8 +15,16 @@ class DrawingViewModel : ViewModel() {
...
+    }
```
brosscode, line 89: extra line detected

app/src/main/java/com/example/drawingApp/DialogUtility.kt
```diff
@@ -0,0 +1,241 @@
...
+        val inflater = LayoutInflater.from(context)
+        val view = inflater.inflate(R.layout.layer_dialog, null)
+        val layerField = view.findViewById<EditText>(R.id.newLayerField)
+        val errorText = view.findViewById<TextView>(R.id.dialogError)
+        val titleText = view.findViewById<TextView>(R.id.layerDialogTitle)
+        if (isCreating) titleText.setText(R.string.new_layer_dialog_title)
+        else titleText.setText(R.string.edit_layer_dialog_title)
+        var layerCounter = 1
+
+
```
brosscode: extra line

app/src/main/java/com/example/drawingApp/DialogUtility.kt
```diff
@@ -0,0 +1,241 @@
...
+import android.widget.Toast
+import androidx.appcompat.app.AlertDialog
+import androidx.core.view.isVisible
+import androidx.core.widget.addTextChangedListener
+import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
+import com.google.android.material.bottomsheet.BottomSheetDialog
+
+//import com.example.drawingApp.R
+
```
brosscode: extra line

app/src/main/java/com/example/drawingApp/DialogUtility.kt
```diff
@@ -0,0 +1,241 @@
...
+import android.widget.TextView
+import android.widget.Toast
+import androidx.appcompat.app.AlertDialog
+import androidx.core.view.isVisible
+import androidx.core.widget.addTextChangedListener
+import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
+import com.google.android.material.bottomsheet.BottomSheetDialog
+
+//import com.example.drawingApp.R
```
brosscode: you have two of these and use neither

app/src/main/java/com/example/drawingApp/EditLayerAdapter.kt
```diff
@@ -0,0 +1,76 @@
+package com.example.drawingApp
+
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.widget.ImageView
+import android.widget.TextView
+import androidx.recyclerview.widget.RecyclerView
+
+
```
brosscode: ex line

app/src/main/java/com/example/drawingApp/EditLayerAdapter.kt
```diff
@@ -0,0 +1,76 @@
...
+class EditLayerAdapter(
+    private val data: List<LayerViewModel>,
```
brosscode, line 11: I did a double take here, but I assume this is a LayerViewModel since you are expecting it to hold a bitmap at some point

Mine is currently a list of strings, but I like the forward thinking.

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -2,20 +2,18 @@ package com.example.drawingApp
 
 import android.os.Bundle
 import android.os.StrictMode
-import android.view.LayoutInflater
-import android.view.View
-import android.widget.EditText
+import android.util.Log
```
brosscode: Log import

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -2,20 +2,18 @@ package com.example.drawingApp
...
 import android.widget.TextView
-import android.widget.Toast
 import androidx.activity.viewModels
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatActivity
-import androidx.core.view.isVisible
-import androidx.core.widget.addTextChangedListener
+import com.google.android.material.bottomsheet.BottomSheetDialog
+
```
brosscode, line 10: extra line

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -26,11 +24,45 @@ class MainActivity : AppCompatActivity() {
         drawingFieldView.onBitmapUpdate = {
             model.activeBitmap = it
         }
-
+        //when we click to create a new layer
```
brosscode: unneeded comment

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -26,11 +24,45 @@ class MainActivity : AppCompatActivity() {
...
         findViewById<TextView>(R.id.newLayer).setOnClickListener {
-            showLayerAlertDialog()
+            //assign the alert dialog to be able to dismiss later
```
brosscode: just gonna leave this review comment here to catch all the comments,

I would review and see what information you think is critical to understanding the reasoning for something, some of these comments are very helpful to knowing what is going on, and others are more of pseudocode for you to follow

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -26,11 +24,45 @@ class MainActivity : AppCompatActivity() {
...
+                wasOpen = model.isLayerCreationDialogOpen,
+                onTextChanged = { model.currentLayerDialogText = it },
+                onDialogSubmission = {
+                    Log.v("Hamdan", "${it}")
+                    if (it != null) {
+                        model.layerNames.add(it)
+                    }
+                    model.isLayerCreationDialogOpen = false
+                },
+                onOpen = { model.isLayerCreationDialogOpen = true }
```
brosscode: is model the reference to the DialogUtility file? or is it the viewModel? or the Adapter LayerViewModel?

my brain is telling me its likely the third one in the list, but this seems like a lot to take in here

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -26,11 +24,45 @@ class MainActivity : AppCompatActivity() {
...
-            showLayerAlertDialog()
+            //assign the alert dialog to be able to dismiss later
+            //show the dialog for new layer
+            alertDialog = DialogUtility.showLayerAlertDialog(
+                context = this,
+                layerNames = model.layerNames,
+                wasOpen = model.isLayerCreationDialogOpen,
+                onTextChanged = { model.currentLayerDialogText = it },
+                onDialogSubmission = {
+                    Log.v("Hamdan", "${it}")
```
brosscode: log

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -26,11 +24,45 @@ class MainActivity : AppCompatActivity() {
...
+            //assign the sheet to be dismissed layer
+            //show the sheet for viewing the layers
+            botSheet = DialogUtility.showViewingLayersDialog(
+                context = this,
+                layerNames = model.layerNames,
+                activeEditIndex = model.layerViewEditDialogIndex,
+                onAlertDialogOpenState = { model.isLayerViewEditDialogOpen = it },
+                onAlertDialogTextChange = { model.currentViewLayerDialogText = it },
+                onEdit = {
+                    model.layerViewEditDialogIndex = it
```
brosscode: why does this one take 3 lines when the ones above it take 1?

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -41,62 +73,35 @@ class MainActivity : AppCompatActivity() {
...
         }
-        //this is here so that in the case that the dialogue was previously open
-        //we will immediately know to set the text to what it was previously
-        //the reason it is here, is so that we avoid overriding the correct text
-        if (model.isLayerCreationDialogOpen) {
-            layerField.setText(model.currentLayerCreationDialogText)
-        }
-        model.isLayerCreationDialogOpen = true
-        dialog.setOnDismissListener { model.isLayerCreationDialogOpen = false }
-        dialog.show()
```
brosscode, line 101: line

app/src/main/java/com/example/drawingApp/DialogUtility.kt
```diff
@@ -0,0 +1,177 @@
+package com.example.drawingApp
+
+//import com.example.drawingApp.R
```
ForzenAcademy: Remove unused comments here

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
HamdanSCode: and again the empty string acts as the effective null

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: nope its the viewmodel because were in the activity, the adapter is only involved when using the adapter.

app/src/main/res/layout/edit_layer_dialog.xml
```diff
@@ -0,0 +1,32 @@
...
+        android:id="@+id/editLayerName"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:layout_weight="3" />
+
+    <ImageView
+        android:id="@+id/layerEditBtn"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:layout_weight="1"
```
ForzenAcademy: These don't need weight. It's standard just to let icons be their normal size. You just need a weight of 1 on the textview.

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -117,7 +137,27 @@ class MainActivity : AppCompatActivity() {
...
+                    layerCounter
+                )
+            )
+        ) {
+            layerCounter += 1
+        }
+        return resources.getString(R.string.new_layer_hint, layerCounter)
     }
+
+//    override fun onDestroy() {    //NEED TO IMPLEMENT THIS TO AVOID WINDOW LEAK
```
ForzenAcademy: Either finish the implementation or let's remove this for now

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -2,101 +2,121 @@ package com.example.drawingApp
...
-        var layerCounter = 1
+            state.activeBitmap?.let {
+                drawingFieldView.setBitmap(it)
+                drawingFieldView.invalidate()
+            }
+            //based on the state of the sheet alertdialog being open or closed as well as if the
+            //sheet is open or closed, change how the submission function of the dialog works
+            if (state.isLayerSheetOpen && state.isAlertDialogOpen) {
+                creating = false
+                onSubmission = {
```
ForzenAcademy: You have enough overlap in these `onSubmissions` you should combine them

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -2,101 +2,121 @@ package com.example.drawingApp
...
-import androidx.core.widget.addTextChangedListener
+
 
 class MainActivity : AppCompatActivity() {
     private val model: DrawingViewModel by viewModels()
+    private var isAlertDialogOpen = false
+    private var isLayerSheetOpen = false
+
+    var botSheetObj: DialogUtility.SheetObject? = null
+    var alertDialog: AlertDialog? = null
```
ForzenAcademy, line 15: If you have an object for both of these, you could probably get rid of your two booleans and just check if these are null or not

app/src/main/java/com/example/drawingApp/EditLayerAdapter.kt
```diff
@@ -0,0 +1,76 @@
...
+class LayerViewHolder(
+    view: View,
+    onEdit: (LayerViewModel) -> Unit,
+    onDelete: (LayerViewModel) -> Unit
+) : ViewHolder(view) {
+    private val textView: TextView
+    private val editView: ImageView
+    private val deleteView: ImageView
+    private val editLayer = onEdit
+    private val deleteLayer = onDelete
```
ForzenAcademy: You're making private variables of things passed in the constructor. No need to do this, just add `val` to the variables in the constructor

app/src/main/java/com/example/drawingApp/EditLayerAdapter.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    override fun onBindViewHolder(holder: ViewHolder, position: Int) {
+        holder.bind(data[position])
+    }
+
+    override fun getItemCount(): Int = data.size
+
+}
+
+class LayerViewModel(val text: String)
```
ForzenAcademy, line 43: Foreshadowing, you might want to make this hold `Layer` instead, but up to you

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
```diff
@@ -7,16 +7,85 @@ import androidx.lifecycle.ViewModel
...
+                activeBitmap,
+                isLayerCreationDialogOpen,
+                currentLayerDialogText,
+                isViewingLayerSheetOpen,
+                layerViewEditDialogIndex
+            )
+        )
+    }
+
+    fun initializeViews() {
```
ForzenAcademy: This is definitely initializing the layer, but is it really initializing a `view`? I could see this as just `initialize` but up to you

app/src/main/java/com/example/drawingApp/DialogUtility.kt
```diff
@@ -0,0 +1,174 @@
...
+
+
+
+
+
+
+
+
+
+
```
ForzenAcademy: So much white space!

app/src/main/res/layout/edit_layer_dialog.xml
HamdanSCode: adjusted

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: removed for now

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: combined

app/src/main/java/com/example/drawingApp/MainActivity.kt
HamdanSCode: removed booleans and implemented null checks instead

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
HamdanSCode: renamed

app/src/main/java/com/example/drawingApp/DialogUtility.kt
HamdanSCode: removed

app/src/main/java/com/example/drawingApp/EditLayerAdapter.kt
HamdanSCode: I will implement this on the ticket its relevant to. We already removed the layer object previously due to it being outside of ticket scope.

app/src/main/java/com/example/drawingApp/EditLayerAdapter.kt
theflag505: Implemented, but I'll be honost I'm not sure how this helps obfuscate this information away.

HamdanSCode: implemented, at first I wasn't sure how this would help obfuscate the information, but I realized I could just remove the variables and just directly use the values passed to the constructor

app/src/main/java/com/example/drawingApp/DialogUtility.kt
```diff
@@ -0,0 +1,157 @@
...
+object DialogUtility {
+
+    const val MAX_CHAR_LEN = 16
+
+    /**
+     * used to updated the sheet adapter with the onEdit lambda
+     * the sheet is for dismissing to avoid window leaks
+     */
+    data class SheetObject(val onEdit: (Int, String) -> Unit, val sheet: BottomSheetDialog)
```
ForzenAcademy: Nit: Lambdas should be the last argument so you can use the special syntax with the {} on the outside

app/src/main/java/com/example/drawingApp/DialogUtility.kt
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * opens a dialog that will take a string and allows you to interact with the string
+     * returns the string to wherever on the onSubmission
+     */
+    fun showLayerAlertDialog(
+        context: Context,
+        startString: String = "",
+        isCreating: Boolean = true,
```
ForzenAcademy: `isCreatingNewLayer`? Hard to look at this and tell what it does and there's no kdoc

app/src/main/java/com/example/drawingApp/DialogUtility.kt
```diff
@@ -0,0 +1,157 @@
...
+            layoutManager = LinearLayoutManager(context)
+            adapter = myAdapter
+        }
+        view.findViewById<TextView>(R.id.closeLayerView).setOnClickListener { sheet.dismiss() }
+
+        sheet.setContentView(view)
+        sheet.setOnDismissListener { onSheetDismiss() }
+        sheet.show()
+
+        return SheetObject(onEdit = { index, str ->
```
ForzenAcademy: This could have the cleaner syntax if you made the suggestion from above

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
ForzenAcademy: I don't think it's good practice to think about if the code will be touched by others. We want to make it good code regardless so we grow into good practices.
I think this field would be better served as nullable, but you can set it up either way obviously. Empty is fine too

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
```diff
@@ -7,16 +7,85 @@ import androidx.lifecycle.ViewModel
...
     var activeBitmap: Bitmap? = null
 
-    var isLayerCreationDialogOpen = false
-    var currentLayerCreationDialogText = ""
+    var isLayerCreationDialogOpen = false   //new Layer
+    var currentLayerDialogText = ""
 
+    var isViewingLayerSheetOpen = false       //edit layer
+
+    //var isLayerViewEditDialogOpen = false   //view inside edit layer
```
ForzenAcademy: Remove unused comment lines

app/src/main/java/com/example/drawingApp/EditLayerAdapter.kt
```diff
@@ -0,0 +1,65 @@
+package com.example.drawingApp
+
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.widget.ImageView
+import android.widget.TextView
+import androidx.recyclerview.widget.RecyclerView
+
+
```
ForzenAcademy: EXTRA. LINE.

ForzenAcademy (approved):
Address comments then merge

app/src/main/java/com/example/drawingApp/DialogUtility.kt
HamdanSCode: isCreating is to tell what the dialog is used for if it's for creating a new layer or editing one. This is purely used to update the title on the dialog.

HamdanSCode: updated the name as well

app/src/main/java/com/example/drawingApp/DialogUtility.kt
HamdanSCode: updated

PR #6, opened 2022-10-07
FA-28 BottomSheet Color picker ready for review

HamdanSCode:
Adjusted and viewstate implemented


app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+         * on null don't do anything with it, don't set rgb and hsv to defaults.
+         */
+        onSubmission: (Color?) -> Unit
+    ) {
+        val sheet = BottomSheetDialog(context)
+        val inflater = LayoutInflater.from(context)
+        val view = inflater.inflate(R.layout.color_picker_bottom_sheet, null)
+        val submit = view.findViewById<TextView>(R.id.dialogSubmit)
+        val cancel = view.findViewById<TextView>(R.id.dialogCancel)
+        val hView = view.findViewById<ForceEditText>(R.id.hslH)
```
brosscode: These views could use _slightly_ more descriptive names. hView is pretty vague, something like hueView would be good enough.

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+        val sView = view.findViewById<ForceEditText>(R.id.hslS)
+        val vView = view.findViewById<ForceEditText>(R.id.hslL)
+        val rView = view.findViewById<ForceEditText>(R.id.rgbR)
+        val gView = view.findViewById<ForceEditText>(R.id.rgbG)
+        val bView = view.findViewById<ForceEditText>(R.id.rgbB)
+        val chView = view.findViewById<ForceEditText>(R.id.colorHex)
+        val slider = view.findViewById<ColorSlider>(R.id.slider)
+        val square = view.findViewById<GradientSquare>(R.id.gradientSquare)
+        val activeBox = view.findViewById<TextView>(R.id.colorOne)
```
brosscode: Very confused on what active box is, this the current color? looks like it to me

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+        val previousBox = view.findViewById<TextView>(R.id.colorTwo)
+        val previousColor = onColorTextForceUpdate().previousColor
+        previousBox.setBackgroundColor(previousColor.toArgb())
+        //used to prevent dragging when clicking on the gradient or the slider
+        val onViewTouch = {
+            sheet.behavior.isDraggable = false
+        }
+        val onViewLift = {
+            sheet.behavior.isDraggable = true
```
brosscode: I like this, swiping things up and down is intuitive on phones, makes sense to toggle this on or off based on touch

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+        }
+        //used to set the color of the active box
+        val setActiveBox: (FloatArray) -> Unit = { hsv ->
```
brosscode: this is a lambda and should start with "on", ex "onSetActiveBox" in this case

brosscode: I'll be pretty honest and say the routing with this one does confuse me a little

so from what I can follow you are passing this lambda as an argument in an invoked lambda (that is called whenever you change something in one of the views). This lambda then updates all of the views in this scope.

I feel like making this a local function could also make this clearer? I am unsure if that approach is better/more professional here, but the way you reset your views feels like it goes through some extra complexity here.

It might be the case where this makes it more reusable? with a sheet like this, I feel like this view is already specialized and improving the readability here should take priority, but Nic will have better insight for this.

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+         * Use it to set wherever you are storing your hue
+         * and then update your color
+         */
+        onHueUpdate: (Float) -> Unit,
+        /**
+         * Called whenever the Saturation or Value is updated via the gradient square
+         * or via text. use it to set wherever you are storing your saturation AND value
+         * update your color afterward
+         */
+        onSatOrValUpdate: (Float, Float) -> Unit,
```
brosscode: No comment but I find it pretty funny we abbreviated these the same way

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+        /**
+         * Called whenever text needs to be updated, meaning on any change, hsv, rgb, or text.
+         * Used to give the information of hsv and previous color.
+         * if there was no previous color pass it black. This is used to help
+         * populate the sheets text views and update the slider and gradient indicators as well as set the previous color box.
+         */
+        onColorTextForceUpdate: () -> DrawingViewModel.ColorPack,
```
brosscode: I tend to prefer using a separate lambda/value to provide a previous value.

I noted photoshop defaults to Red on startup, but that's just a random preference, any color should be fine.

Also, I asked Nic about it, but you might want to consider putting the data class in a ColorUtil file and instead only holding an array or hex string in the viewModel.

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+        /**
+         * Called whenever one of the RGB texts is changed by focus by a user.
+         * it passes a ColorInt as the first parameter, used to decide what value to set.
+         * the second int is the actual value you will set your R,G,or B value to.
+         * Color.RED corresponds to red,Color.GREEN to green, BLUE to blue
+         */
+        onRGBTextChange: (Int, Int) -> Unit,
```
brosscode: I was able to handle this locally in the view and didn't use a lambda for it. I would consider handling it that way and only using lambdas to provide information to the viewModel to hold onto.

For example, I only currently have one Lambda in the dialog, and it is to provide the hex string to the viewModel, and on instantiation in the onUpdate, I give it the current hex string in the viewModel, and default that string to #ff0000

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+        /**
+         * called whenever the dialog is dismissed or submitted
+         * on actual submission the color will not be null so take it in and set it wherever you need the selected color and set the hsv and rgb fields to defaults.
+         * on null don't do anything with it, don't set rgb and hsv to defaults.
+         */
+        onSubmission: (Color?) -> Unit
```
brosscode: for a color gradient I wouldn't reset the hsv and rgb to defaults, a user might expect to see those remain at the values they were next time they enter the dialog, since they would be on their current color, but this is a designer question to be sure.

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+            sheet.dismiss()
+        }
+
+        slider.onHueChange = {
+            square.hue = it
+            square.invalidate()
+            onHueUpdate(it)
+            updateForceViews()
+        }
+        square.onSatOrValChange = { saturation, value ->
```
brosscode: your going to want to release the focus you have on your editTexts here, if they are in focus they won't update properly from what I encountered.

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+                    )
+                ).toColor()
+            )
+            sheet.dismiss()
+        }
+        cancel.setOnClickListener {
+            sheet.dismiss()
+        }
+
+        slider.onHueChange = {
```
brosscode: recommend clearing focus here

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+        square.onSatOrValChange = { saturation, value ->
+            onSatOrValUpdate(saturation, value)
+            square.sat = saturation
+            square.valu = value
+            updateForceViews()
+        }
+        square.onDown = onViewTouch
+        square.onUp = onViewLift
+        slider.onDown = onViewTouch
+        slider.onUp = onViewLift
```
brosscode: 👍 cool implementation of this logic

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+
+        hView.addTextChangedListener {
+            val passedInt =
+                hView.text.toString().toFloatOrNull()?.let { string ->
```
brosscode: I put this in a color Util function and just had it return itself as a boolean, then I used an if statement to validate that the input was good. We should probably ask Nic how we should handle incorrect inputs in the edit text, some applications I have used for drawing before would ignore the letters and update just based on the colors in there.

Most likely the answer will be to not update until the error is corrected, but you never know for sure.

To be clear, I don't think you need to change this if it is working for you

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+            }
+        }
+        sView.addTextChangedListener {
+            val saturation = sView.text.toString().toFloatOrNull()
+                ?.let { string -> max(min(1f, string.div(100)), 0f) }
+            val value = vView.text.toString().toFloatOrNull()
+                ?.let { string -> max(min(1f, string.div(100)), 0f) }
+            saturation?.let { sat ->
+                value?.let { valu ->
+                    onSatOrValUpdate(sat, valu)
```
brosscode: good idea to ignore the hue here, I didn't do that and was inefficient and was assigning a hue value when I didn't need to do it

👍

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,102 +1,77 @@
...
+                    onSatOrValUpdate = { saturation, value ->
+                        model.hsvS = saturation
+                        model.hsvV = value
+                        model.updateColorFromHSV()
+                        model.updateRGBFromColor()
+                    },
+                    onColorTextForceUpdate = {
+                        model.getPack()
+                    },
+                    onRGBTextChange = { color, colorVal ->
```
brosscode: love the when block here

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,102 +1,77 @@
...
+                            }
+                        }
+                        model.updateHSVFromRGB()
+                        model.updateColorFromHSV()
+                    },
+                    onSubmission = { color ->
+                        val isNull = color == null
+                        color?.let {
+                            model.chosenColor = it
+                            Toast.makeText(
```
brosscode: 🍞

app/src/main/res/layout/activity_main.xml
```diff
@@ -36,9 +36,24 @@
...
             android:text="@string/dialogPopUpBtn"
             android:textSize="@dimen/text_1" />
+        <TextView
+            android:id="@+id/colorPickerBtn"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:background="@drawable/text_btn_background_"
+            android:padding="@dimen/grid_1"
+            android:text="@string/dialogPopUpBtn"
+            android:textSize="@dimen/text_1" />
```
brosscode, line 54: extra line, I'll only put this once so you can sweep these when you are at the point where that polish is meaningful

app/src/main/java/com/example/drawingApp/MainActivity.kt
```diff
@@ -1,102 +1,77 @@
...
 import android.widget.TextView
 import android.widget.Toast
 import androidx.activity.viewModels
-import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatActivity
-import androidx.core.view.isVisible
-import androidx.core.widget.addTextChangedListener
 
 class MainActivity : AppCompatActivity() {
     private val model: DrawingViewModel by viewModels()
```
brosscode, line 17: consider renaming to viewModel

app/src/main/java/com/example/drawingApp/GradientSquare.kt
```diff
@@ -0,0 +1,172 @@
...
+        const val imageSize = 175
+
+        var bitmaps: Array<Bitmap?>? = Array(360) { null }
+        var completedIndexes: Int? = 0
+        var isCaching: Boolean? = false
+        var areBitmapsDone: Boolean? = false
+        /**
+         * used in the case that the static values somehow gets deleted and set to null
+         */
+        fun regenerate() {
```
brosscode: 💸

app/src/main/java/com/example/drawingApp/GradientSquare.kt
```diff
@@ -0,0 +1,172 @@
...
+                    }
+                }
+            }
+            getBitmap()?.let { c.drawBitmap(it, null, drawRect, null) }
+            c.drawCircle(sat * width, height - valu * height, 5f, selectionPaint)
+        }
+    }
+
+    companion object {
+        const val imageSize = 175
```
brosscode: we can go SMALLER

brosscode (commented):
Lots of good stuff in here! let me know if you got questions about any of the comments or if something I left was a little vague!

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
HamdanSCode: It already does so.

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
HamdanSCode: already does so.

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
HamdanSCode: I think this implementation is fine, because it will only take something in the valid range of input because everything gets sanitized to be in range, and if its not able to be taken as a float its taken as null and in that case nothing happens.

app/src/main/java/com/example/drawingApp/GradientSquare.kt
HamdanSCode: we can, but I started noticing visual differences at 150 and smaller looks worse.

HamdanSCode: 175 is already really fast

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
HamdanSCode: it's the current color thats actively shown and edited by the user.

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
HamdanSCode: this lambda is esssentialy a local function, rather than giving a function signature and having to pass the function the views and all that it seemed better to just acknowledge this was only going to be used in this scope and to just create it as a lambda to be used in the function.

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
HamdanSCode: putting the data class in the util file does make sense

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
HamdanSCode: fair enough

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+import android.graphics.Color
+import android.view.LayoutInflater
+import android.widget.TextView
+import androidx.core.graphics.toColor
+import androidx.core.widget.addTextChangedListener
+import com.google.android.material.bottomsheet.BottomSheetDialog
+import kotlin.math.max
+import kotlin.math.min
+
+object ColorPickerUtility {
```
ForzenAcademy: `Util` is much more common

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
HamdanSCode: adjusted to activeColorBox

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+
+    @SuppressLint("ClickableViewAccessibility")
+    fun colorPickerSheet(
+        context: Context,
+        /**
+         * Called whenever the hue is updated via text or slider.
+         * Use it to set wherever you are storing your hue
+         * and then update your color
+         */
+        onHueUpdate: (Float) -> Unit,
```
ForzenAcademy: Why do you need both these lambdas? Why not just a "onColorUpdate"? Does anyone outside the scope of this sheet need individual updates?

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+            onHueUpdate(it)
+            updateForceViews()
+        }
+        square.onSatOrValChange = { saturation, value ->
+            onSatOrValUpdate(saturation, value)
+            square.sat = saturation
+            square.valu = value
+            updateForceViews()
+        }
+        square.onDown = onViewTouch
```
ForzenAcademy: These are oneliners, don't need these lambdas

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+            val value = vView.text.toString().toFloatOrNull()
+                ?.let { string -> max(min(1f, string.div(100)), 0f) }
+            saturation?.let { sat ->
+                value?.let { valu ->
+                    onSatOrValUpdate(sat, valu)
+                    updateForceViews()
+                }
+            }
+        }
+        rView.addTextChangedListener {
```
ForzenAcademy: Do you have these set in your layout to only allow numeric input?

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+            }
+        }
+        bView.addTextChangedListener {
+            val blue = bView.text.toString().toIntOrNull()?.let { it1 -> min(max(it1, 0), 255) }
+            blue?.let {
+                onRGBTextChange(Color.BLUE, it)
+                updateForceViews()
+            }
+        }
+        chView.addTextChangedListener {
```
ForzenAcademy: `hexView`?

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+            val hex = chView.text.toString()
```
ForzenAcademy: You're not expecting the # to be in the text field?

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
```diff
@@ -0,0 +1,222 @@
...
+        rView.addTextChangedListener {
+            val red = rView.text.toString().toIntOrNull()?.let { it1 -> min(max(it1, 0), 255) }
+            red?.let {
+                onRGBTextChange(Color.RED, it)
+                updateForceViews()
+            }
+        }
+        gView.addTextChangedListener {
+            val green = gView.text.toString().toIntOrNull()?.let { it1 -> min(max(it1, 0), 255) }
+            green?.let {
```
ForzenAcademy: For these, don't store a `green` etc. Just do `?.let {` on the big calculation
`gView.text.toString().toIntOrNull()?.let { it1 -> min(max(it1, 0), 255) }?.let {`

app/src/main/java/com/example/drawingApp/ColorSlider.kt
```diff
@@ -0,0 +1,125 @@
...
+
+@SuppressLint("ClickableViewAccessibility")
+class ColorSlider @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyle: Int = 0
+) : View(context, attrs, defStyle) {
+
+    private var held = false
+    var hue: Float = 180f
```
ForzenAcademy: 180F should be companion constant

app/src/main/java/com/example/drawingApp/ColorSlider.kt
```diff
@@ -0,0 +1,125 @@
...
+
+    private fun getBitmap(): Bitmap? {
+        return if (sliderBitmap == null) {
+            sliderBitmap = generateBitmap()
+            sliderBitmap
+        } else sliderBitmap
+    }
+
+    private fun generateBitmap(): Bitmap {
+        var startHue = 360f
```
ForzenAcademy: 360 should be constant

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
HamdanSCode: removed and incorported an onColorUpdate

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
HamdanSCode: adjusted

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
HamdanSCode: I did not, now they are incorporated

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
HamdanSCode: incorporated in the force text function

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
HamdanSCode: updated the name

app/src/main/java/com/example/drawingApp/ColorPickerUtility.kt
HamdanSCode: no I am not, it is on the Prompt side

app/src/main/java/com/example/drawingApp/ColorSlider.kt
HamdanSCode: fair enough, also incorporated a min hue constant

app/src/main/res/layout/color_picker_bottom_sheet.xml
```diff
@@ -0,0 +1,201 @@
...
+                android:text="B:" />
+
+            <com.example.drawingApp.ForceEditText
+                android:id="@+id/rgbB"
+                style="@style/Widget.AppCompat.EditText"
+                android:layout_width="50dp"
+                android:layout_height="40dp"
+                android:inputType="number"
+                android:text="0"
+                android:textSize="10dp" />
```
ForzenAcademy: Text sizes are sp

app/src/main/res/layout/color_picker_bottom_sheet.xml
```diff
@@ -0,0 +1,201 @@
...
+                android:text="G:" />
+
+            <com.example.drawingApp.ForceEditText
+                android:id="@+id/rgbG"
+                style="@style/Widget.AppCompat.EditText"
+                android:layout_width="50dp"
+                android:layout_height="40dp"
+                android:inputType="number"
+                android:text="0"
+                android:textSize="10dp" />
```
ForzenAcademy: Same same, all over

app/src/main/res/layout/color_picker_bottom_sheet.xml
```diff
@@ -0,0 +1,201 @@
...
+
+        <LinearLayout
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:orientation="horizontal">
+
+            <TextView
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:text="G:" />
```
ForzenAcademy: String resources for everything

app/src/main/res/layout/color_picker_bottom_sheet.xml
```diff
@@ -0,0 +1,201 @@
...
+
+            <TextView
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:text="R:" />
+
+            <com.example.drawingApp.ForceEditText
+                android:id="@+id/rgbR"
+                style="@style/Widget.AppCompat.EditText"
+                android:layout_width="50dp"
```
ForzenAcademy: Constraining the size of text widgets is a no bueno. Accessibility wise this could make them cut off text

app/src/main/res/layout/color_picker_bottom_sheet.xml
```diff
@@ -0,0 +1,201 @@
...
+            android:layout_height="wrap_content"
+            android:orientation="horizontal">
+
+            <TextView
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:text="L:" />
+
+            <com.example.drawingApp.ForceEditText
+                android:id="@+id/hslL"
```
ForzenAcademy: NO MORE L. V.

app/src/main/res/layout/color_picker_bottom_sheet.xml
```diff
@@ -0,0 +1,201 @@
...
+
+        <LinearLayout
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:orientation="horizontal">
+
+            <TextView
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:text="L:" />
```
ForzenAcademy: NO

app/src/main/java/com/example/drawingApp/GradientSquare.kt
```diff
@@ -0,0 +1,176 @@
...
+        var completedIndexes: Int? = 0
+        var isCaching: Boolean? = false
+        var areBitmapsDone: Boolean? = false
+
+        /**
+         * used in the case that the static values somehow gets deleted and set to null
+         */
+        fun regenerate() {
+            if (bitmaps == null) {
+                bitmaps = Array(360) { null }
```
ForzenAcademy: Constant

app/src/main/java/com/example/drawingApp/GradientSquare.kt
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+
+    companion object {
+        const val imageSize = 175
+
+        var bitmaps: Array<Bitmap?>? = Array(360) { null }
+        var completedIndexes: Int? = 0
+        var isCaching: Boolean? = false
+        var areBitmapsDone: Boolean? = false
```
ForzenAcademy: Privates

app/src/main/java/com/example/drawingApp/GradientSquare.kt
```diff
@@ -0,0 +1,176 @@
...
+            c.drawCircle(sat * width, height - valu * height, 5f, selectionPaint)
+        }
+    }
+
+    companion object {
+        const val imageSize = 175
+
+        var bitmaps: Array<Bitmap?>? = Array(360) { null }
+        var completedIndexes: Int? = 0
+        var isCaching: Boolean? = false
```
ForzenAcademy: What are the differences between these two booleans?

ForzenAcademy: So maybe I wasn't super clear, on statics you can lose their value, which will reset them. For objects this means they will go back to null. For primitives they will go to their default value. So for these primitives just make them not null

app/src/main/java/com/example/drawingApp/GradientSquare.kt
```diff
@@ -0,0 +1,176 @@
...
+        }
+        createdBitmap.setPixels(pixels, 0, imageSize, 0, 0, imageSize, imageSize)
+        return createdBitmap
+    }
+
+    @SuppressLint("DrawAllocation")
+    override fun onDraw(canvas: Canvas?) {
+        super.onDraw(canvas)
+        canvas?.let { c ->
+            if (!::drawRect.isInitialized) drawRect = Rect(0, 0, width, height)
```
ForzenAcademy: Let's avoid lateinit var for this, no need

app/src/main/java/com/example/drawingApp/GradientSquare.kt
```diff
@@ -0,0 +1,176 @@
...
+            synchronized(lock) {
```
ForzenAcademy: 👍

app/src/main/java/com/example/drawingApp/GradientSquare.kt
```diff
@@ -0,0 +1,176 @@
...
+                        completedIndexes = completedIndexes?.plus(1)
+                    }
+                    areBitmapsDone = true
+                    isCaching = false
+                }
+            }
+        }
+    }
+
+    private fun getBitmap(): Bitmap? {
```
ForzenAcademy: Why nullable? Are you ever returning null here?

app/src/main/java/com/example/drawingApp/GradientSquare.kt
```diff
@@ -0,0 +1,176 @@
...
+        val rIndex = max(0, min(hue.toInt(), 359))
```
ForzenAcademy: Constant.

Also 360 is the same as 0 so there's a slight bit of rounding logic that can occur here forcing it back to 359 if it's say "359.9" 🤷

app/src/main/java/com/example/drawingApp/GradientSquare.kt
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+
+    private fun startCacheTask() {
+        GlobalScope.launch(Dispatchers.IO) {
+            regenerate()
+            completedIndexes?.let {
+                val startPoint = completedIndexes
+                if (startPoint != null) {
+                    for (i in startPoint..359) {
```
ForzenAcademy: Use a constant for 360

Use an exclusive range to not go till 360. `until`?

app/src/main/java/com/example/drawingApp/GradientSquare.kt
```diff
@@ -0,0 +1,176 @@
...
+                        regenerate()
+                        bitmaps?.set(i, generateBitmap(i))
+                        completedIndexes = completedIndexes?.plus(1)
+                    }
+                    areBitmapsDone = true
```
ForzenAcademy: What are the difference of these?

app/src/main/java/com/example/drawingApp/GradientSquare.kt
```diff
@@ -0,0 +1,176 @@
...
+    init {
+        setOnTouchListener { _, motionEvent ->
+            sat = max(min(motionEvent.x / width, 1f), 0f)
+            valu = max(min((height - motionEvent.y) / height, 1f), 0f)
+            when (motionEvent.action) {
+                MotionEvent.ACTION_DOWN -> {
+                    held = true
+                    onDown?.invoke()
+                    (this.parent as? ViewGroup)?.children?.forEach { it.clearFocus() }
+                    onSatOrValChange(sat, valu)
```
ForzenAcademy: Change these to `s` and `v` or use their full name. I don't pray to `Valu`

app/src/main/java/com/example/drawingApp/GradientSquare.kt
```diff
@@ -0,0 +1,176 @@
...
+     * used to prevent dragging
+     */
+    var onDown: (() -> Unit)? = null
+
+    /**
+     * called when user lifts off gradient or cancels,
+     * allows the user to drag dialog again
+     */
+    var onUp: (() -> Unit)? = null
+    var sat = 1f
```
ForzenAcademy: Privates?

app/src/main/java/com/example/drawingApp/ForceEditText.kt
```diff
@@ -0,0 +1,43 @@
...
+        textListeners.forEach { addTextChangedListener(it) }
+    }
+
+    fun forceText(s: String) {
+        removeListeners()
+        if (!this.hasFocus()) setText(s)
+        addBackListeners()
+    }
+
+    val rgbInt: Int?
```
ForzenAcademy: 👍

app/src/main/java/com/example/drawingApp/ForceEditText.kt
```diff
@@ -0,0 +1,43 @@
...
+        textListeners.forEach { removeTextChangedListener(it) }
+    }
+
+    fun addBackListeners() {
+        textListeners.forEach { addTextChangedListener(it) }
+    }
+
+    fun forceText(s: String) {
+        removeListeners()
+        if (!this.hasFocus()) setText(s)
```
ForzenAcademy: Why not wrap this whole function in this if statement? Why remove listeners and then put them back if you're not changing text?

app/src/main/java/com/example/drawingApp/ForceEditText.kt
```diff
@@ -0,0 +1,43 @@
...
+import kotlin.math.max
+import kotlin.math.min
+
+class ForceEditText @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyle: Int = 0
+) : EditText(context, attrs, defStyle) {
+
+    private var textListeners = mutableListOf<TextWatcher>()
```
ForzenAcademy: This can be `val`

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
```diff
@@ -1,22 +1,58 @@
...
-
+    data class ViewState(
+        val isColorSheetOpen: Boolean
+    )
+
+    private var isColorPickerSheetOpen = false
+
+    var chosenColor: Color? = null
+
+    var hsvH = 0f
```
ForzenAcademy: 🤔 `h`? lol
Really, use your `Hsv` data class for this

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
```diff
@@ -1,22 +1,58 @@
...
+    var hsvS = 0f
+    var hsvV = 0f
+
+    //used to help determine rgb values for consistency
+    var activeColor: Color = Color.BLACK.toColor()
```
ForzenAcademy: Return an Int for your colors. This is a bit messy using the object. No need. Colors are just numbers

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
```diff
@@ -1,22 +1,58 @@
...
-    var currentLayerCreationDialogText = ""
-
-
+    data class ViewState(
+        val isColorSheetOpen: Boolean
+    )
+
+    private var isColorPickerSheetOpen = false
+
+    var chosenColor: Color? = null
```
ForzenAcademy: `Int`

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
```diff
@@ -1,22 +1,58 @@
 package com.example.drawingApp
```
ForzenAcademy: Is this a second viewmodel now?

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
```diff
@@ -1,22 +1,58 @@
...
+
+    fun updateViewState() {
+        onUpdate?.invoke(
+            ViewState(
+                isColorSheetOpen = isColorPickerSheetOpen
+            )
+        )
+    }
+
+    fun updateColorFromHSV() {
```
ForzenAcademy: Acronyms are still camel case

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
```diff
@@ -1,22 +1,58 @@
...
+        val color = HSVToColor(floatArrayOf(hsvH, hsvS, hsvV))
```
ForzenAcademy: You can make a method in your Hsv data class for toColor even

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
```diff
@@ -1,22 +1,58 @@
...
-    val layerNames: MutableList<String> = mutableListOf()
-
-    var activeBitmap: Bitmap? = null
-
-    var isLayerCreationDialogOpen = false
-    var currentLayerCreationDialogText = ""
-
-
+    data class ViewState(
+        val isColorSheetOpen: Boolean
```
ForzenAcademy: At least put the currently chosen color in here too just to make a bit more fleshed out. Could help in rotation?

app/src/main/java/com/example/drawingApp/ColorSlider.kt
```diff
@@ -0,0 +1,127 @@
...
+            getBitmap()?.let { c.drawBitmap(it, 0f, 0f, null) }
+            val lineY = (MAX_HUE - min(hue, MAX_HUE)) / MAX_HUE * height
+            c.drawLine(0f, lineY, width.toFloat(), lineY, selectionPaint)
+        }
+    }
+
+    companion object {
+        const val MIN_HUE = 0f
+        const val MAX_HUE = 360f
+        var sliderBitmap: Bitmap? = null
```
ForzenAcademy: Private?

app/src/main/java/com/example/drawingApp/ColorSlider.kt
```diff
@@ -0,0 +1,127 @@
...
+
+    private val selectionPaint = Paint().apply {
+        strokeWidth = 5f
+        style = Paint.Style.STROKE
+        color = Color.WHITE
+    }
+
+    private fun getBitmap(): Bitmap? {
+        return if (sliderBitmap == null) {
+            sliderBitmap = generateBitmap()
```
ForzenAcademy: Maybe
`return generateBitmap().also { sliderBitmap = it }` ? Or something like that

app/src/main/java/com/example/drawingApp/ColorSlider.kt
```diff
@@ -0,0 +1,127 @@
...
+        invalidate()
+    }
+
+    private val selectionPaint = Paint().apply {
+        strokeWidth = 5f
+        style = Paint.Style.STROKE
+        color = Color.WHITE
+    }
+
+    private fun getBitmap(): Bitmap? {
```
ForzenAcademy: Why return nullable? Is there any branch where this can return null?

Looking at this more, why not something like

```
private fun getBitmap() = sliderBitmap ?: generateBitmap().also { sliderBitmap = it }
```

ForzenAcademy: 😎

app/src/main/java/com/example/drawingApp/ColorSlider.kt
```diff
@@ -0,0 +1,127 @@
...
+                        onHueChange?.invoke(hueCoordinate)
+                    }
+                }
+            }
+            true
+        }
+    }
+
+    fun sliderSetHue(newHue: Float) {
+        hue = newHue
```
ForzenAcademy: This isn't triggering onHueChange. Intended??

app/src/main/java/com/example/drawingApp/ColorSlider.kt
```diff
@@ -0,0 +1,127 @@
...
+        invalidate()
+    }
+
+    private val selectionPaint = Paint().apply {
```
ForzenAcademy: Put in Companion?

app/src/main/java/com/example/drawingApp/ColorSlider.kt
```diff
@@ -0,0 +1,127 @@
...
+import kotlin.math.min
+
+@SuppressLint("ClickableViewAccessibility")
+class ColorSlider @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyle: Int = 0
+) : View(context, attrs, defStyle) {
+
+    private var held = false
```
ForzenAcademy: Put this below the others so it's more separated

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
```diff
@@ -0,0 +1,201 @@
...
+        colorHexView.addTextChangedListener {
+            val hex = colorHexView.text.toString()
+            try {
+                val color = Color.parseColor("#$hex")
+                val hsv = FloatArray(3)
+                Color.colorToHSV(color, hsv)
+                slider.sliderSetHue(hsv[0])
+                onColorUpdate(Hsv(hsv[0], hsv[1], hsv[2]))
+                updateForceViews()
+            } catch (_: Exception) {
```
ForzenAcademy: 1. Ctrl + S
2. At least LOG it lol

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
```diff
@@ -0,0 +1,201 @@
...
+            val value = valueView.text.toString().toFloatOrNull()
+                ?.let { string -> max(min(1f, string.div(100)), 0f) }
+            value?.let {
+                onColorUpdate(Hsv(value = value))
+                updateForceViews()
+            }
+        }
+        val onRgbTextChange = {
+            if (rgbRedView.rgbInt != null && rgbGreenView.rgbInt != null && rgbBlueView.rgbInt != null) {
+                val hsv = rgbToHsv(rgbRedView.rgbInt!!, rgbGreenView.rgbInt!!, rgbBlueView.rgbInt!!)
```
ForzenAcademy:
```
rgbToHsv(rgbRedView.rgbInt!!, rgbGreenView.rgbInt!!, rgbBlueView.rgbInt!!).let {
onColorUpdate(Hsv(it[0], it[1], it[2]))
}
```

No need to store a variable for a one off like that

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
```diff
@@ -0,0 +1,201 @@
...
+        //used to prevent dragging while interacting with either gradient or slider
+        square.onDown = { sheet.behavior.isDraggable = false }
+        square.onUp = { sheet.behavior.isDraggable = true }
+        slider.onDown = { sheet.behavior.isDraggable = false }
+        slider.onUp = { sheet.behavior.isDraggable = true }
+
+        hueView.addTextChangedListener {
+            val hue = hueView.text.toString().toFloatOrNull()
+                ?.let { string ->
+                    max(min(360f, string), 0f)
```
ForzenAcademy: Constant

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
```diff
@@ -0,0 +1,201 @@
...
+                }
+            hue?.let {
+                onColorUpdate(Hsv(hue = hue))
+                updateForceViews()
+            }
+        }
+        saturationView.addTextChangedListener {
+            val saturation = saturationView.text.toString().toFloatOrNull()
+                ?.let { string -> max(min(1f, string.div(100)), 0f) }
```
ForzenAcademy: Use math operators instead of arithmetic functions `/ 100`

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
```diff
@@ -0,0 +1,201 @@
...
+        }
+        previousColorBox.setOnClickListener {
+            val hsv = FloatArray(3)
+            Color.colorToHSV(previousColor.toArgb(), hsv)
+            //used to set the color of the active box to the previous color
+            onSettingActiveBox(hsv)
+        }
+        submit.setOnClickListener {
+            onSubmission(
+                Color.HSVToColor(
```
ForzenAcademy: Use your fancy Hsv object with the method you're gonna make for this

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
```diff
@@ -0,0 +1,201 @@
...
+            activeColorBox.setBackgroundColor(intColor)
+        }
+        //called whenever text needs to be updated
+        val updateForceViews = {
+            val pack = onColorTextForceUpdate()
+            onSettingActiveBox(floatArrayOf(pack.hue, pack.saturation, pack.value))
+        }
+        previousColorBox.setOnClickListener {
+            val hsv = FloatArray(3)
+            Color.colorToHSV(previousColor.toArgb(), hsv)
```
ForzenAcademy: Hsv object?

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
```diff
@@ -0,0 +1,201 @@
...
+
+            square.hue = hsv[0]
+            square.sat = hsv[1]
+            square.valu = hsv[2]
+            square.invalidate()
+
+            slider.sliderSetHue(hsv[0])
+
+            activeColorBox.setBackgroundColor(intColor)
+        }
```
ForzenAcademy: Space after this

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
```diff
@@ -0,0 +1,201 @@
...
+        val rgbBlueView = view.findViewById<ForceEditText>(R.id.rgbB)
+        val colorHexView = view.findViewById<ForceEditText>(R.id.colorHex)
+        val slider = view.findViewById<ColorSlider>(R.id.slider)
+        val square = view.findViewById<GradientSquare>(R.id.gradientSquare)
+        val activeColorBox = view.findViewById<TextView>(R.id.colorOne)
+        val previousColorBox = view.findViewById<TextView>(R.id.colorTwo)
+        val previousColor = onColorTextForceUpdate().previousColor
+        previousColorBox.setBackgroundColor(previousColor.toArgb())
+        //used to set the color of the active box
+        val onSettingActiveBox: (FloatArray) -> Unit = { hsv ->
```
ForzenAcademy: Take in Hsv object

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
```diff
@@ -0,0 +1,201 @@
...
+import kotlin.math.max
+import kotlin.math.min
+
+object ColorPickerUtil {
+
+    data class ColorPack(
+        val hue: Float,
+        val saturation: Float,
+        val value: Float,
+        val previousColor: Color,
```
ForzenAcademy: This is silly you have an Hsv object lol

app/src/main/res/layout/color_picker_bottom_sheet.xml
HamdanSCode: fair enough, if theres any design changes you want made please tell as well

app/src/main/java/com/example/drawingApp/GradientSquare.kt
HamdanSCode: isCaching: to tell if were in the process of creating bitmaps, areBitmapsDone: to tell if we have finished making bitmaps.

HamdanSCode: removed the optional from them.

app/src/main/java/com/example/drawingApp/GradientSquare.kt
HamdanSCode: no but a thread could suddenly make a bitmap null as were about to return. In reality, we would never end up returning a null though because we aren't using any threads that would turn a value to null.
I added 2 double bangs to allow it to return without being angry.

app/src/main/java/com/example/drawingApp/GradientSquare.kt
HamdanSCode: thats fine though to have that happen right?
added the constant.

app/src/main/java/com/example/drawingApp/GradientSquare.kt
HamdanSCode: implemented

app/src/main/java/com/example/drawingApp/GradientSquare.kt
HamdanSCode: I see what you mean.

app/src/main/java/com/example/drawingApp/GradientSquare.kt
HamdanSCode: none for the purposes of use, removed one

app/src/main/java/com/example/drawingApp/GradientSquare.kt
HamdanSCode: they are referred and used in the dialog sheet, If I make getters and setters I could make these private, but the getters and setters kind of make it not private anyway.

app/src/main/java/com/example/drawingApp/ForceEditText.kt
HamdanSCode: fair point, no point in doing extra nothing

app/src/main/java/com/example/drawingApp/ForceEditText.kt
HamdanSCode: your right, its a reference to the starting point of the list, I can still change the values in the list.

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
HamdanSCode: who should the hsv data class belong to?
I'm almost thinking it should be its own standalone data class at this point rather than sitting in the util or the viewmodel.

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
HamdanSCode: your right, it did feel very awkward doing it the way I was.

HamdanSCode: implemented as Ints

app/src/main/java/com/example/drawingApp/GradientSquare.kt
HamdanSCode: Actualy, just realized this is how I'm making sure not to call cache() over and over while not finished, we need to know to continue cacheing if we close and reopen the coroutine as well (in the case I have to implement that later).

app/src/main/java/com/example/drawingApp/GradientSquare.kt
HamdanSCode: Actually there is a specific line it is important, to tell whether we need to start cacheing, because while in the process of cacheing we are not done, if I only have one bool I'll constantly start cacheing over and over.

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
HamdanSCode: I'm sorry I dont know what you mean? Theres only one android view model.

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
HamdanSCode: fair enough, it would be needed for when the user is actually using it, wouldn't hurt to throw it in the view state.

app/src/main/java/com/example/drawingApp/ColorSlider.kt
HamdanSCode: fair, implemented

app/src/main/java/com/example/drawingApp/ColorSlider.kt
HamdanSCode: won't this generate the bitmap everytime though? The intention with the other way was to make it so that if we already have it ready in the static just use it.

app/src/main/java/com/example/drawingApp/ColorSlider.kt
HamdanSCode: well, that definitely is a much cleaner implementation, it feels good that I actually understand what it says too

app/src/main/java/com/example/drawingApp/ColorSlider.kt
HamdanSCode: the second implementation offered is much cleaner.

app/src/main/java/com/example/drawingApp/ColorSlider.kt
HamdanSCode: It previously caused a loop and crashed, it no longer does so I implemented it.

app/src/main/java/com/example/drawingApp/ColorSlider.kt
HamdanSCode: fair, it is consistent.

app/src/main/java/com/example/drawingApp/ColorSlider.kt
HamdanSCode: implemented I think?

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
HamdanSCode: fair enough, also realized I'm only calling the function in one place, no real need to store that either

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
HamdanSCode: created

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
HamdanSCode: implemented

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
HamdanSCode: implemented

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
HamdanSCode: implemented

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
HamdanSCode: implemented

app/src/main/java/com/example/drawingApp/GradientSquare.kt
HamdanSCode: Implemented as an enum, however the issue potentially still persists that if the job stops, we will say we are cacheing when the process is no longer cacheing.

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
```diff
@@ -0,0 +1,209 @@
...
+        val previousColor: Int
+    )
+
+    fun rgbToHsv(red: Int, green: Int, blue: Int): FloatArray {
+        val hsv = FloatArray(3)
+        Color.RGBToHSV(red, green, blue, hsv)
+        return hsv
+    }
+
+//    fun colorToHsv(color: Int): Hsv {
```
ForzenAcademy: Remove unused sections

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
```diff
@@ -0,0 +1,209 @@
...
+         */
+        onColorTextForceUpdate: () -> ColorPack,
+        /**
+         * called whenever the dialog is dismissed or submitted
+         * on actual submission the color will not be null so take it in and set it wherever you need the selected color and set the hsv and rgb fields to defaults.
+         * on null don't do anything with it, don't set rgb and hsv to defaults.
+         */
+        onSubmission: (Int?) -> Unit
+    ) {
+        val MAX_HUE = 360f
```
ForzenAcademy: Don't make these local to a function, make them in the object

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
```diff
@@ -0,0 +1,209 @@
...
+
+        //called whenever text needs to be updated
+        val updateForceViews = {
+            val pack = onColorTextForceUpdate()
+            onSettingActiveBox(pack.hsv)
+        }
+
+        previousColorBox.setOnClickListener {
+            //used to set the color of the active box to the previous color, impossible to be null here
+            onSettingActiveBox(Hsv.colorToHsv(previousColor))
```
ForzenAcademy: Your Hsv object already makes this code so much nicer to read

app/src/main/java/com/example/drawingApp/ColorSlider.kt
```diff
@@ -0,0 +1,119 @@
...
+            true
+        }
+    }
+
+    fun sliderSetHue(newHue: Float) {
+        hue = newHue
+        invalidate()
+    }
+
+    private fun getBitmap() = sliderBitmap ?: generateBitmap().also { sliderBitmap = it }
```
ForzenAcademy: 😍

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
```diff
@@ -1,22 +1,66 @@
...
+            ViewState(
+                isColorSheetOpen = isColorPickerSheetOpen,
+                chosenColor = chosenColor ?: Color.BLACK
+            )
+        )
+    }
+
+    fun updateColorFromHsv() {
+        val color = hsv.toColor()
+        activeColor = color!!
```
ForzenAcademy: 1. Why make a color variable just to set it a line later?
2. No need to double bang this

`activeColor = hsv.toColor() ?: activeColor` ?

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
```diff
@@ -1,22 +1,66 @@
...
+
+    fun initialize() {
+        updateViewState()
+    }
+
+    fun setHsv(newHsv: Hsv) {
+        hsv = newHsv
+    }
+
+    fun getHsv(): Hsv {
```
ForzenAcademy: `fun getHsv() = hsv`

app/src/main/java/com/example/drawingApp/ForceEditText.kt
```diff
@@ -0,0 +1,45 @@
...
+            removeListeners()
+            setText(s)
+            addBackListeners()
+        }
+    }
+
+    val rgbInt: Int?
+        get() {
+            return this.text.toString().toIntOrNull()
+                ?.let { intValue -> min(max(intValue, 0), 255) }
```
ForzenAcademy: Constant

app/src/main/java/com/example/drawingApp/GradientSquare.kt
```diff
@@ -0,0 +1,169 @@
...
+
+@SuppressLint("ClickableViewAccessibility")
+class GradientSquare @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyle: Int = 0
+) : View(context, attrs, defStyle) {
+
+    private val lock: Any = Any()
+    private val selectionPaint = Paint().apply {
```
ForzenAcademy: Companion

app/src/main/java/com/example/drawingApp/Hsv.kt
```diff
@@ -0,0 +1,31 @@
+package com.example.drawingApp
+
+import android.graphics.Color
+
+data class Hsv(
+    val hue: Float? = null,
```
ForzenAcademy: Why make these nullable? You don't need to return a nullable color if these aren't nullable. If any of these would be null, make the whole object null, not one of the fields. It seems odd to have an object that represents a color but then someone wouldn't provide all the properties of the color fields HSV.

app/src/main/java/com/example/drawingApp/Hsv.kt
```diff
@@ -0,0 +1,31 @@
...
+                saturation!!,
+                value!!
+            )
+        )
+        else null
+    }
+
+    companion object{
+
+        fun colorToHsv(color: Int): Hsv {
```
ForzenAcademy: Nice static method. Works well here

app/src/main/res/layout/color_picker_bottom_sheet.xml
```diff
@@ -0,0 +1,208 @@
...
+        <TextView
+            android:id="@+id/colorOne"
+            android:layout_width="match_parent"
+            android:layout_height="35dp"
+            android:background="@color/black" />
+
+        <TextView
+            android:id="@+id/colorTwo"
+            android:layout_width="match_parent"
+            android:layout_height="35dp"
```
ForzenAcademy, line 34: Don't give height to the textviews in here, you may cut off text

app/src/main/res/values/dimens.xml
```diff
@@ -26,4 +32,10 @@
     <dimen name="corner_rounding_5">18dp</dimen>
     <dimen name="corner_rounding_6">20dp</dimen>
 
+    <dimen name="editText_1">40dp</dimen>
```
ForzenAcademy, line 35: What are these for?

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
HamdanSCode: removed

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
HamdanSCode: fair enough, I tried to see if I could put const in the function and when it said no I just moved on and made it a val, I'll keep this in mind for the future

app/src/main/java/com/example/drawingApp/ColorSlider.kt
HamdanSCode: lmao

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
HamdanSCode: good point, remade it without a double bang.
I think the reason why it was like that previously was that the function did more previously and with pruning stuff was removed from it.

app/src/main/java/com/example/drawingApp/DrawingViewModel.kt
HamdanSCode: god sometimes this syntax is cracked

app/src/main/java/com/example/drawingApp/ForceEditText.kt
HamdanSCode: created

app/src/main/java/com/example/drawingApp/GradientSquare.kt
HamdanSCode: moved to companion

app/src/main/res/layout/color_picker_bottom_sheet.xml
HamdanSCode: swapped to wrap

app/src/main/res/values/dimens.xml
HamdanSCode: If in the future we want to implement editText fields or larger editTextFields

app/src/main/java/com/example/drawingApp/Hsv.kt
HamdanSCode: thank you

app/src/main/java/com/example/drawingApp/Hsv.kt
```diff
@@ -0,0 +1,22 @@
...
+data class Hsv(
+    val hue: Float,
+    val saturation: Float,
+    val value: Float,
+) {
+    fun toColor() = Color.HSVToColor(floatArrayOf(hue, saturation, value))
+
+    companion object {
+
+        fun colorToHsv(color: Int): Hsv {
```
ForzenAcademy: `fromColor` would be more standard here

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
```diff
@@ -0,0 +1,202 @@
...
+object ColorPickerUtil {
+    const val MAX_HUE = 360f
+    const val SAT_VAL_FACTOR = 100
+
+    data class ColorPack(
+        val hsv: Hsv,
+        val previousColor: Int
+    )
+
+    fun rgbToHsv(red: Int, green: Int, blue: Int): FloatArray {
```
ForzenAcademy: Returning an HSV here would be nicer I think

ForzenAcademy (approved):

app/src/main/java/com/example/drawingApp/Hsv.kt
HamdanSCode: implemented

app/src/main/java/com/example/drawingApp/ColorPickerUtil.kt
HamdanSCode: implemented

